PR #14, opened 2021-11-18
Reorganize volvelle top wheel

apoelstra:
Fixes #6 

Fixes #5

PR #15, opened 2021-11-18
Underline 6 and 9 on both volvelle wheels

### LINKED ISSUE #12: The 6 and 9 characters are somewhat ambigous on the wheel.
* One idea is to place the characters above the arrow on the wheel to help disambiguate them. Or put the characters under the box with an upward arrow.
* A more forceful solution is to change fonts for the 9 character.

apoelstra:
Builds on #14 but is a separate PR because there's more room for bikeshedding (e.g. should we underline other characters? all the characters? try a different approach)

Fixes #12

### DIFF
--- a/SSS32.ps
+++ b/SSS32.ps
@@ -185,7 +185,13 @@
    [ exch 1 exch dup { 32 idiv exch } forall 0 exch { 31 and } forall ] polymod0
  } bind def
 
-/centreshow {dup stringwidth pop 2 div neg 0 rmoveto show} bind def
+/underlineshow {
+    dup dup (6) eq exch (9) eq or { % if the string is (6) or (9)
+        gsave (_) show grestore % draw an underline
+    } if
+    show
+} bind def
+/centreshow {dup stringwidth pop 2 div neg 0 rmoveto underlineshow} bind def
 /centresquare {dup neg 2 div dup rmoveto dup 0 rlineto dup 0 exch rlineto neg 0 rlineto closepath stroke} bind def
 /concatstrings % (a) (b) -> (ab)
    { exch dup length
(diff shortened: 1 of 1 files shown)


roconnor-blockstream:
After reviewing 12 sided dice on etsy, The common solutions here are underlining or followed by a period.

If we want to do a period we need some tighter kerning than offered natively by the Courier font.

To draw an underline, I recommend the simpler code of `gsave (_) show grestore`.
To draw a dot, the following code seems to work okay, but feel free to modify it.

    gsave ( ) stringwidth pop 2 div 0 rmoveto (.) show grestore

While I like the underline better in general, I worry that it may not show up on the bottom disc if people have ill-cut out or ill-fastened to discs.  I think using a period is more robust.

apoelstra:
Rebased and changed to your underline code. That's much simpler!

I will experiment with the period ... but my feeling, after cutting out an underlined disk imperfectly, is that the underline is okay. Generally if you mess up the wheel then either the top or bottom of characters is consistently cut off ... and if you cut off the bottom, this is like a synthetic underline, while if you cut off the top then the printed underline is visible.

If you're only a little bit off, I actually usually force the wheel a little bit until I can see the underline, just to be sure.

I'm also a little skeptical that a `.` is any less likely to be hidden by a poorly-cut wheel than an underline.

I may also just try both.

apoelstra:
I tried a few variants (just on my screen, not printing and cutting) and I think I continue to prefer the underline.

Just the dot looks a little awkward, especially as it feels poorly-placed no matter where you put it along the body of the 6. I'm not sure that its purpose is obvious. And the dot + underline is too noisy.

roconnor-blockstream:
One other issue with underline to be aware of is that there is a third option to distinguish 6 and 9, and that is to use an overline. :/

apoelstra:
One thing that helps guide this is that with the new top wheel layout from #14, the numbers 2-9 appear in order, so it is clear from context which one is 6 and which one is 9 ... and therefore whether the line is an underline or an overline.

roconnor-blockstream:
Okay, let's try the underline.  Some comments I've read on dice says that the underline is more distinct, and I've never actually encountered anyone trying to use an overline.

PR #19, opened 2021-11-29
WIP: block out recovery instruction page; add giant 3-of-n recovery symbol table

apoelstra:
The table spans 5 pages and is labelled "Appendix B" but it is not yet moved to the end of the document, nor is the set of creation tables labelled "Appendix A". Will do this in a later PR.

This is a WIP for now because the instructions are incomplete and also I don't like the way that they flow. This does not include the recovery share table nor instructions for how to use it, it just suggests using the volvelle. And the instructions feel overly general which hides how simple the process is for k = 2.

### DIFF
--- a/SSS32.ps
+++ b/SSS32.ps
@@ -1,6 +1,6 @@
 %!PS-Adobe-3.0
 %%Orientation: Portrait
-%%Pages: 21
+%%Pages: 27
 %%EndComments
 %%BeginSetup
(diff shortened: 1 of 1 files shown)


apoelstra:
Closing this because I no longer think the giant table is worthwhile. For 3-of-n it's super fast to just use the new recovery slide rule and multiplication slide rule.

PR #20, opened 2021-11-30
Change share generation page to start from A rather than S

apoelstra:
This currently adds an extra page with "Constructing from an existing secret" instructions. I think this is OK to merge but in the future we may want to move this extra page to a separate module.

There is also a fair bit of duplicated code between the two "Constructing Shares" pages but it's not clear to me how to factor it out.


roconnor-blockstream:
Probably best to entirely remove the S column.  If someone wants to recover their secret from their random shares, they can use the recovery disc.

apoelstra:
I agree. Done.

apoelstra:
Closing this since this page will be replaced when we do the general content rendering.

PR #21, opened 2021-11-30
WIP: addition worksheet

apoelstra:
This is a concept for an addition worksheet. It has 3 sections which each let you add up to 5 full-length strings together.

Lots of parameters to mess with here, I'm not sure how to approach this in the most generally useful way.

### DIFF
--- a/SSS32.ps
+++ b/SSS32.ps
@@ -987,20 +987,23 @@ showpage
   } bind def
 
   /x 96 def
-  /y 660 def
+  /y 670 def
 
-  3 { %repeat
+  5 { %repeat n sections
     x y (ms1) drawRow
     /y y 14 sub def
-    4 { %repeat
-      x y (  +) drawRow
-      /y y 2 sub def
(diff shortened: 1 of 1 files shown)


apoelstra:
Added an alternate design that I like better, that has 5 sections, each of which let you add up to 3 things together

apoelstra:
Added an alternate design which is just 15 pairs of rows, which is probably the most generally useful

roconnor-blockstream:
Unfortunately the operation that needs to be done is multiply a share by a symbol, and add that value to an accumulated result, and repeat.

For 2-of-n creation, a separate worksheet is needed for people using the tables, where people just fill in their all their n shares at once one column (depending on how you want to orient things) at a time.

apoelstra:
In my usage I translate symbols by reading them off the checksum worksheet, translating them, and writing the result in the "Addition worksheet". So the translation step I can do in "one line".

apoelstra:
Closing. Will PR again after the new addition worksheet is in.

PR #23, opened 2021-12-21
clean up preamble; add title page

apoelstra:
The first commit can be reviewed with `git show --color-moved=zebra`; you will see there is no red (removed) text and the only green (added) text is comments. The rest are moves.

This also adds a dedicated title page, mainly so I could propose a block-comment format for pages.


roconnor-blockstream:
A variant of this was merged in 10a771481f0edbf22c60f817f5d1705d21654d70.

PR #27, opened 2021-12-24
Replace "dice table / multiplication" page with table page

apoelstra:
The dice table will be replaced by a more robust layout that Russell has in mind.


apoelstra:
Rebased and squashed everything.

danai554 (commented):
Danaikhamkao

PR #29, opened 2021-12-29
add reference page

apoelstra:
Adds a sheet with the data/header format, binary conversions and a table giving suggested pronunciations of all the symbols.

### DIFF
--- a/SSS32.ps
+++ b/SSS32.ps
@@ -1,6 +1,6 @@
 %!PS-Adobe-3.0
 %%Orientation: Portrait
-%%Pages: 23
+%%Pages: 24
 %%EndComments
 %%BeginSetup
 [(Shamir's Secret) (Sharing Codex)]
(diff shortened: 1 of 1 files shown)


apoelstra:
Ready for review.

roconnor-blockstream:
```
  /aleph       (Aleph)      /alpha    (Alpha)    /beta        (Beta)           /Gamma     (Gamma)
  /Delta       (Delta)      /epsilon  (Epsilon)  /eta         (Eta)            /Theta     (Theta)
  /Lambda      (Lambda)     /mu       (Mu)       /Xi          (Xi)             /Pi        (Pi)
  /rho         (Rho)        /Sigma    (Sigma)    /Phi         (Phi)            /Psi       (Psi)
  /Omega       (Omega)      /at       (At)       /numbersign  (Hash)           /percent   (Percent)
  /cent        (Cent)       /yen      (Yen)      /Euro        (Euro)           /currency  (Scarab)
  /circleplus  (Earth)      /dagger   (Dagger)   /daggerdbl   (Double-dagger)  /section   (Section)
  /paragraph   (Paragraph)  /diamond  (Diamond)  /heart       (Heart)
```

This is my proposed revised pronunciation table.

apoelstra:
concept ACK your new table, though I am getting postscript errors when I try to use it.

apoelstra:
Ah - the reason is that my logic assumes that `/multiply` is present, but your table removes it. Ok, I'll fix my logic.

apoelstra:
@roconnor-blockstream updated with your new table

roconnor-blockstream:
Probably due to the rebase, but the number of pages needs to be increased and the last two pages need to be renumbered properly.

roconnor-blockstream:
functions like `/position` `/drawBrace` need to be moved out of the page drawing section and put in the `SETUP` section.

`/boxW` `/pronunciation`, etc. all need to be bound to a disposable dictionary (i.e. surround by `10 dict begin ... end` or also moved out.

roconnor-blockstream:
Slightly prefer `2NAMEAXXXXXXXXXXXXXXXXXXXXXXXXXXCE43R337JKVTZ` to be used as the example.

apoelstra:
Will fix the numbering and example string. Will also refactor the code to just move all the drawing stuff into the separate functions of SETUP section and use `translate` and stack parameters for positioning etc

apoelstra:
@roconnor-blockstream completely refactored it -- now the three sections have their own drawing logic in functions in the main setup section; the drawing code then just uses `translate` to position things, draws the titles, and calls these funcitons.

roconnor-blockstream:
Codex no longer lives as a font and must be rendered with a `codexshow` or other fuction

roconnor-blockstream:
I think we preferred the binary tables to be transposed.

apoelstra:
Updated to use `codexshow`; transposed the tables.

PR #32, opened 2022-03-27
add translation worksheet

apoelstra:


apoelstra:
Closing, since this is in conflict wtih master, and the conflicts will get worse before they get better. Will re-introduce as part of a wider push to redo the whole document.

PR #34, opened 2022-03-31
Add basic test vectors

robot-dreams:


apoelstra:
Thanks! These look great, but I think we want to include the HRP everywhere, because in particular we use `MS` for 128-bit secrets but I think `MN` for 256-bit ones. (Is that right @roconnor-blockstream ?)

apoelstra:
I guess we should also decide on a hrp for 512 bit secrets..

roconnor-blockstream:
HRP will be the same.  It will just contain two codewords.

roconnor-blockstream:
One last request:  Can you lowercase most, but not all, of the examples?  Lowercase is canonical.  We use upper case in the volvelles because upper case letters are more clear for printing.

robot-dreams:
Yeah good idea, kept Test Vector 2 as uppercase since it's the most likely volvelle use case.

apoelstra:
Looks great to me -- though I haven't yet verified the vectors.

robot-dreams:
@apoelstra Thanks! You can verify them by checking / running this script and comparing the diffs: https://gist.github.com/robot-dreams/436cc0ff7f244d96510389a8b0165ea7

(The script was NOT optimized for efficiency or readability, haha, but hopefully it gets the point across.)

roconnor-blockstream:
I have verified the test vectors with an independent implementation.

PR #39, opened 2022-05-10
General suggestions for BIP cleanup

robot-dreams:


apoelstra:
This looks great to me! I think @roconnor-blockstream may take issue with the "such as" language in the first paragraph, since I think this scheme is supposed to be used *specifically* with BIP-32 seeds.

robot-dreams:
@apoelstra sounds good, updated to make it clear it's specifically for BIP-32

Added a bit more copy editing. One thing to note, previous we had this terminology:

- An MS32 string with share index != s is called an **MS32 share**
- An MS32 string with share index = s is called an **MS32 secret**
- The component of the "data part" that actually contains the Bech32 encoded data is called a **secret share**

I think this could cause some confusion—in the PR I renamed the last thing to **payload** instead, but LMK if you prefer a different name.

apoelstra:
Thanks, this all looks great to me!

PR #40, opened 2022-05-23
Swap reference/tables pages; convert many fonts to Times

apoelstra:
Should be non-controversial. Purely stylistic changes in anticipation of adding a lot more text content.


apoelstra:
Closing. Will recreate this once #52 is in, since then page reordering is much simpler to verify (and to do).

PR #43, opened 2023-02-13
First pass at BIP improvements

apoelstra:
Changes Python code to use 4-space tabs; renames and reorders some sections; adds BIP header, Rationale and Backward Compatibility sections.

What needs to be done now is to go over all the existing text, which has been compressed into the Specification section, and break it up a bit.

### DIFF
--- a/MasterSeed32.md
+++ b/MasterSeed32.md
@@ -1,10 +1,25 @@
-# Master Seed 32
+# codex32
 
 WARNING: Seriously, this is a work in progress, and it is only a concept right now.
 If you try to use this for your valuable data, I promise you will lose your data.
 You will lose this document and come back here only to find that I have made incompatible changes,
 and your data is lost forever. Even if you don't lose this document, there is no warranty or
 guarantee of any kind that you will be able to successfully recover your data.
+
+```
+  BIP: ????
+  Layer: Applications
+  Title: codex32
+  Author: <list of authors' real names and email addrs>
+  Comments-URI: https://github.com/bitcoin/bips/wiki/Comments:BIP-????
+  Status: Draft
+  Type: <Standards Track | Informational | Process>
+  Created: 2023-02-13
+  License: BSD-3-Clause
+* Post-History: <dates of postings to bitcoin mailing list, or link to thread in mailing list archive>
+```
+
+# Introduction
 
 ## Abstract
 
@@ -14,6 +29,11 @@ It includes an encoding format, a BCH error-correcting checksum, and algorithms 
 Secret data can be split into up to 31 shares.
 A minimum threshold of shares, which can be between 1 and 9, is needed to recover the secret,
 whereas without sufficient shares, no information about the secret is recoverable.
+
+## Copyright
+
+This document is licensed under the 3-clause BSD license.
+
 
 ## Motivation
 
@@ -48,7 +68,7 @@ Note that hand computation is optional,
 the particular details of hand computation are outside the scope of this standard,
 and implementers do not need to be concerned with this possibility.
 
-## Specification
+# Specification
 
 ### MS32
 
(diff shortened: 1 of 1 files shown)


MasterSeed32.md
```diff
@@ -274,9 +309,32 @@ Generation of long shares and recovery of the master seed from long shares proce
 The long checksum is designed to be an error correcting code that can correct up to 4 character substitutions, up to 8 unreadable characters (called erasures), or up to 15 consecutive erasures.
 As with regular checksums we do not specify how an implementation should implement error correction, and all our recommendations for error correction of regular MS32 strings also apply to long MS32 strings.
 
-## Test Vectors
+# Rationale
+
+Storing master secrets is the most important component of self-custody.
```
roconnor-blockstream, line 314: master seeds

MasterSeed32.md
```diff
@@ -274,9 +309,32 @@ Generation of long shares and recovery of the master seed from long shares proce
...
+
+* Use BIP39, which suffers from a weak checksum, uses English words which are have inconsistent lengths and near collisions (or other word lists, the choice of which affects the derived secret but is not explicitly encoded anywhere), and has no natural way to add Shamir Secret Sharing; or
+* Use SLIP39, which also uses English words, includes features such as passphrase-based hardening that may not be worth the additional implementation complexity, and does not have a BIP number.
+
+codex32 is a simpler version of SLIP39 which has a slightly stronger checksum, a more compact and language-independent encoding, and is simple enough that all parts of the scheme (including error detection but not error correction), can be implemented without the use of electronic computers.
+
+# Backwards Compatibility
+
+codex32 is an alternative to BIP39 and SLIP39.
+It is technically possible, but not supported by this BIP, for users to re-encode their BIP39 or SLIP39 secrets in the codex32 format, and to re-encode back.
```
roconnor-blockstream, line 326: I'm inclined to strike "but not supported by this BIP".

MasterSeed32.md
```diff
@@ -274,9 +309,32 @@ Generation of long shares and recovery of the master seed from long shares proce
...
+However, the only supported way for users to move between codex32 and other formats is to generate a new seed and to sweep their coins.
```
roconnor-blockstream, line 327: I think this needs maybe a bit of work to clarify.

1. It is possible to move master seeds from BIP39 and SLIP39 to Codex32.
2. It is not possible to move shares from SLIP39 to Codex32.
3. It is not possible to move master seeds from Codex32 to BIP39.
4. It is not recommended to move master seeds from BIP39 to Codex32 as it requires storing 512-bit shares and secrets.

MasterSeed32.md
```diff
@@ -1,11 +1,26 @@
...
 and your data is lost forever. Even if you don't lose this document, there is no warranty or
 guarantee of any kind that you will be able to successfully recover your data.
 
+```
+  BIP: ????
+  Layer: Applications
+  Title: codex32
+  Author: <list of authors' real names and email addrs>
+  Comments-URI: https://github.com/bitcoin/bips/wiki/Comments:BIP-????
+  Status: Proposed
```
roconnor-blockstream, line 15: Maybe start with draft until the analysis of the generators is completed?

MasterSeed32.md
apoelstra: I mainly put it in there to avoid questions about "if it's possible then how do you do it?".

I'll drop it for now.

MasterSeed32.md
roconnor-blockstream: I admit I got this wrong in several places.

MasterSeed32.md
roconnor-blockstream: Actually I think BIP2 requires a full implementation somewhere before we can move to Proposed, and the postscript doesn't implement the long strings.

MasterSeed32.md
```diff
@@ -274,9 +309,38 @@ Generation of long shares and recovery of the master seed from long shares proce
...
+
+codex32 is an alternative to BIP39 and SLIP39.
+It is technically  possible to derive the BIP32 master seed from seed words encoded in one of these schemes, and then to encode this seed in codex32.
+For BIP39 this process is irreversible, since it involves hashing the original words.
+Furthermore, the resulting seed will be 512 bits long, which may be too large to be safely and conveniently handled.
+
+SLIP39 seed words can be reversibly converted to master seeds, so it is possible to interconvert between SLIP39 and codex32.
+However, SLIP39 **shares** cannot be converted to codex32 shares because the two schemes use a different underlying field.
+
+However, the authors of this BIP do not recommend interconversion. Instead, users who wish to switch to codex32 should generate a fresh seed and sweep their coins.
```
roconnor-blockstream, line 333: Maybe drop the leading "However"

MasterSeed32.md
```diff
@@ -150,40 +176,43 @@ In order to recover a master seed, one needs a set of valid MS32 shares such tha
...
-    res ^= a if ((b >> i) & 1) else 0
-    a *= 2
-    if (32 <= a):
-      a ^= 41
-  return res
+    res = 0
+    for i in range(5):
+        res ^= a if ((b >> i) & 1) else 0
+        a *= 2
+        if (32 <= a):
```
roconnor-blockstream, line 189: Do you want to make this

    a ^= 41 if (32 <= a) else 0

to match the style above?

Maybe for another PR?

MasterSeed32.md
```diff
@@ -274,9 +309,38 @@ Generation of long shares and recovery of the master seed from long shares proce
...
 
-## Test Vectors
+# Rationale
+
+Storing master seeds is the most important component of self-custody.
+Users who do not wish to use multisignature setups or sweep their coins must do so by encoding the master secret and physically encoding it.
+The current best practices for this are either to:
+
+* Use BIP39, which suffers from a weak checksum, uses English words which are have inconsistent lengths and near collisions (or other word lists, the choice of which affects the derived secret but is not explicitly encoded anywhere), and has no natural way to add Shamir Secret Sharing; or
+* Use SLIP39, which also uses English words, includes features such as passphrase-based hardening that may not be worth the additional implementation complexity, and does not have a BIP number.
```
roconnor-blockstream, line 319: SLIP-39 (should be hyphenated I think) also has a two level sharing heirarchy that we do not support (in this BIP).

suggest "...includes features such as passphrase-hardening and two level sharing that may not..."

MasterSeed32.md
apoelstra: It says that a lack of reference implementation won't block getting a number, so no need to prioritize this .. but I'll put together a full Python impl from the existing reference code and sipa's pychar stuff.

MasterSeed32.md
apoelstra: I think we should leave this as-is. I'm not sure if `<assignment> if <condition> else 0` is valid Python but I think it's surprising to have a side-effecting operation in an if expression.

roconnor-blockstream: You are parsing it wrong it is `<value> if <condition> else 0` and we do it two lines above.

But I'm fine to consider it in a separate PR.

apoelstra: Oh! I see. Yeah, I like that suggestion. But we'll do it in a separate PR.

PR #52, opened 2023-03-18
add default.nix which can produce ps and pdf files

apoelstra:
Will also patch the git commit into the "revision" variable which appears in the ps output, if told to build a git repo rather than the current source tree.


default.nix
```diff
@@ -0,0 +1,218 @@
...
+  # Checks whether the generated output matches the checked-in SSS32.ps.
+  # When doing development you may want to shut this off to obtain the
+  # output file that you need to check in.
+, doOutputDiff ? true
+}:
+
+let
+  src =
+    if isNull ref
+    then ./.
```
roconnor-blockstream, line 16: Probably want some sort of filter source here.  I find that `./.` gets really polluted, especially with `result` symlinks from builds.

default.nix
```diff
@@ -0,0 +1,218 @@
...
+    finalContent = builtins.foldl' addPage initialContent booklet.pages;
+  in
+    pkgs.writeTextFile {
+      name = booklet.name;
+      text = finalContent.content + ''
+        %%EOF
+      '';
+    };
+in
+stdenv.mkDerivation {
```
roconnor-blockstream, line 200: Why not just put renderBooklet as the derivation result?  Wouldn't that make it easier to check the result into git?  We can do `nix-build -o SSS32.ps` and then commit SSS32.ps. (I haven't actually tried this).

apoelstra: Because I eventually want to produce more than one output -- the full booklet, but also postscript files with individual worksheets, 256-bit vesrions of the worksheets, etc. And maybe a illustrated version alongside the regular version.

default.nix
apoelstra: Sure -- I suspect we actually only want the .ps and .inc files, so this should be an easy filter.

apoelstra: Added a commit which filters the source.

apoelstra:
Force pushed to a .nix which more closely matches the file I'm actually working with to add the full booklet contents, color, etc.

You can see how the dependency management is going to work, once we break up the setup code into more than a single file, and we output the full booklet, but it's hopefully clearer how we'd output other documents too.

roconnor-blockstream:
99463184ca456853d5624e3830d439fd58ddd933 seems to add an extra blank line on line 2060 for no good reason.

default.nix
```diff
@@ -0,0 +1,247 @@
...
+    ghostscriptTest() {
+      echo "Ghostscript testing $1"
+      local output;
+      if ! output="$(gs -dNOPAUSE -dNODISPLAY "$1" < /dev/null 2>&1)"; then
+        echo "Failed to run ghostscript on $1"
+        echo "$output"
+        exit 1
+      fi
+    }
+  '' + lib.optionalString doPageChecks toString (
```
roconnor-blockstream, line 225: Should this all be in a standard `checkPhase` gated by a standard `doCheck` flag?

default.nix
```diff
@@ -0,0 +1,222 @@
...
+    then lib.sourceFilesBySuffices ./. [ ".ps" ".inc" ]
+    else
+      fetchGit {
+        url = ./.;
+        inherit ref;
+      };
+  shortId = lib.optionalString (! isNull ref) ("-" + builtins.substring 0 8 src.rev);
+
+
+  setup = rec {
```
roconnor-blockstream, line 28: probably best to remove `rec`.

default.nix
apoelstra: Ah, yes, it should be.

default.nix
```diff
@@ -0,0 +1,222 @@
...
+          else
+            if pageData ? isLandscape
+            then "90 rotate\n"
+            else ""
+        ) + ''
+          ${pageData.content}
+          ${lib.optionalString (pageData ? drawPageContent) "end\n"}pgsave restore
+          showpage
+        '';
+        nextCtIdx = content.nextCtIdx + (if pageData ? drawPageContent then 1 else 0);
```
roconnor-blockstream, line 174: Should `CtIdx` just be `PgIdx` - 1, or is that just a coincidence?

apoelstra:
@roconnor-blockstream I'll check on the "no good reason" blank line while in flight today, but I suspect that it's there because Nix multiline strings always end in a newline (or is it that they start with a newline? I forget), so if we don't have these newlines in the target output, it forces you to cram lots of stuff onto single lines.

default.nix
apoelstra: It's just a coincidence, but let me check my current development branch to see if we want this calculation at all. It originates with the PostScript originally having a giant array of page content in the setup section, but now that we're assembling pages in Nix, I'm going to move the content into their respective pages rather than indexing into an array.

Regarding the page numbering logic, it's going to wind up being a little ugly, since we need to support

* Pages that get numbers
* Pages that are numbered, but the number isn't displayed because it'd interfere with the content (at least one volvelle page are like this)
* Pages that aren't numbered at all

I'd also like to support some preamble pages having numbers like `i`, `ii`, `iii`, etc.

Anyway for now maybe I just want to rip out the page numbering logic entirely since we need to iterate on it.

default.nix
apoelstra: On further reflection, and having learned Nix better, let me explore making this derivation have multiple outputs, where the default is the fully-assembled SSS32.ps. I'm unsure if this is the right way to go still, because there are e.g. flags we want to be able to set about PDF vs PS, color vs non-color etc, and it might still be the cleanest thing to output a directory with a bunch of different files.

roconnor-blockstream: My experiments with git and symlinks was a failure.  While we could support multiple outputs, I do not think that it worth worrying about.  At least not at this time.

default.nix
apoelstra: I can remove it for now, but it will be used -- this `setup` set will contain all the various components of the postscript "setup" section. Each will have a `dependencies` array which refers to other setup components.

apoelstra:
@roconnor-blockstream pushed new version which

* Renames `nextCtIdx` to `nextFooterIdx` to match my local "complete" branch and to reflect that this value indicates the page number to be drawn in the footer, not an index into any content
* Removes the extraneous blank line :)
* Splits the page tests and diff into a `checkPhase` section and added a `doCheck` flag
* Minor refactorings

default.nix
```diff
@@ -0,0 +1,251 @@
...
+    let
+      addPage = content: pageData: {
+        content = content.content + lib.optionalString (pageData ? sourceHeader) ''
+          %****************************************************************
+          %*
+          %* ${pageData.sourceHeader}
+          %*
+          %****************************************************************
+        '' + ''
+          %%Page: ${toString content.nextPgIdx} ${toString content.nextPgIdx}
```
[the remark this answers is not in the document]
apoelstra: @roconnor-blockstream yep, I'll address that in a future PR. Unfortunately `ps2pdf` loses this information (or maybe PDF doesn't support it) so it's of somewhat limited value.

default.nix
```diff
@@ -0,0 +1,221 @@
...
+
+  buildInputs = if doPdfGeneration then [ ghostscript ] else [ ];
+
+  phases = [ "buildPhase" ];
+  buildPhase = ''
+    set -e
+
+    mkdir "$out"
+    cd "$out"
+    cp ${renderBooklet fullBooklet} SSS32.ps
```
roconnor-blockstream: Should this be a symlink instead?

default.nix
```diff
@@ -0,0 +1,221 @@
...
+
+    ${lib.optionalString doOutputDiff "diff -C 5 ${src}/SSS32.ps SSS32.ps"}
+    sed -i 's/(revision \(.*\))/(revision \1${shortId})/' ./SSS32.ps
```
roconnor-blockstream, line 237: This probably needs to be more robust, but it is fine for now.

default.nix
```diff
@@ -0,0 +1,221 @@
...
+      name = booklet.name;
+      text = finalContent.content + ''
+        %%EOF
+      '';
+    };
+in
+stdenv.mkDerivation {
+  name = "codex32${shortId}";
+
+  buildInputs = if doPdfGeneration then [ ghostscript ] else [ ];
```
roconnor-blockstream: I think this ought to be `nativeBuildInputs`.

default.nix
```diff
@@ -199,23 +204,50 @@ let
...
-  buildInputs = if doPdfGeneration then [ ghostscript ] else [ ];
+  buildInputs = if doCheck || doPdfGeneration then [ ghostscript ] else [ ];
+
+  phases = [ "buildPhase" ] ++ lib.optionals doCheck [ "checkPhase" ];
 
-  phases = [ "buildPhase" ];
   buildPhase = ''
     set -e
 
+    ghostscriptTest() {
```
roconnor-blockstream, line 220: This would make more sense inside `checkPhase`.

default.nix
apoelstra: Hmm, yeah, I guess it could be. It didn't occur to me.

default.nix
apoelstra: Ok, will change. I really don't understand the distinction.

apoelstra:
Rebased to address commetns.

default.nix
roconnor-blockstream: My vague understanding is that `buildInputs` is historic, and means build and runtime dependencies.  While `nativeBuildInputs` is just buildtime dependencies.  I'm not sure what the actual effect of this difference is.

default.nix
```diff
@@ -0,0 +1,251 @@
...
+
+  buildPhase = ''
+    set -e
+
+    FULL_BW="${renderBooklet fullBooklet}"
+
+    # Copy output Postscript into place
+    mkdir "$out"
+    cd "$out"
+    ln -s "$FULL_BW" SSS32.ps
```
roconnor-blockstream: You changed this from `cp` to `ln -s`, but below you patch the file with `sed`.  Does this even work?  the content of `FULL_BW` should be an immutable nix-store file, so I don't think you can patch it.

roconnor-blockstream: Should renderBooklet take a revision argument?

apoelstra: No, it does not work. The `sed` line appears to be doing nothing, and since it does nothing, does not even try to write to the file, and returns success.

As for "should renderBooklet just take a `revision` argument?" let me investigate that. Would be more elegant than rendering the book and then patching it after-the-fact, but might make templating a bit weird.

apoelstra:
Will undraft when this is ready for review again. Hopefully today or tomorrow.

default.nix
apoelstra: After testing this by setting the `ref` parameter, I believe the existing code *does* work. I am investigating how. It's as though the `ln -s` is doing a copy-on-write copy rather than a softlink. (I can confirm that the target file is modified by `sed` but the original file is not.)

apoelstra: Oh, `sed -i` will replace linked files unless you pass it `--follow-symlinks`.

Linux.

apoelstra:
Restored the old `cp` line rather than using `ln -s` and relying on `sed`'s magic link-replacing behavior.

Added Makefile which figures out the git commit ID and passes it to nix.

VzxPLnHqr:
@roconnor-blockstream @apoelstra 
Apologies to barge in on this PR, but it seemed the most relevant location given that you are nixifying codex32 things here and that would be immensely helpful. 

I have a (currently failed) [attempt](https://github.com/VzxPLnHqr/nixos-airgapped/compare/static-codex32?expand=1) to add [apoelstra/volvelle-website](https://github.com/apoelstra/volvelle-website) to [nixos-airgapped](https://github.com/VzxPLnHqr/nixos-airgapped). My attempt is failed currently because there is some sort of permission error when trying to run `wasm-pack` and my rust/nix skills are not the greatest yet. You may be able to diagnose in 30 seconds and it would be greatly appreciated!

Also, is that volvelle-website repo even the correct one to be doing this with? I noticed that this repo seems to have more activity. Do you intend to add some sort of `nix build .#codex32-offline-website` magic to this repo that builds the static website?

apoelstra:
>You may be able to diagnose in 30 seconds and it would be greatly appreciated!

Afraid not. I just checked my local nix CI scripts to see if I got wasm-pack working, and the answer is no, I only have wasm-pack code in one place, and it's all commented out with a confusing explanation. Unfortunately wasm-pack is a nightmare to use in an airgapped context, because it downloads random crapand  does not respect lockfiles.

> Also, is that volvelle-website repo even the correct one to be doing this with? I noticed that this repo seems to have more activity.

Yeah, this repo is where the "real" work is happening. The website is more-or-less a demo and an aid for doing workshops. (I think I will do one in October at TABconf so probably you'll see a bunch of activity there in the week(s) leading up to that, and nothing until then.)

But you raise an interesting point that we should have a self-contained "codex32 tool" that people can build and use locally. WASM is nice because you can run it in the browser and easily have nice interactive interfaces, but it's pretty awful in every other respect.

One soluion for you would be to just build the site and nix-package the binary output....that'd fly in Nix land even though it wouldn't fly in Guix. But I definitely understand if you refuse on principle to do that!

>Do you intend to add some sort of `nix build .#codex32-offline-website` magic to this repo that builds the static website?

Probably not, unless we can get away from wasm-pack. (Which honestly shouldn't be too hard ... in theory I just need to run cargo with a wasm32-unknown-unknown target and then write a tiny bit of glue code ... but I don't know how.)

VzxPLnHqr:
> ...see if I got wasm-pack working, and the answer is no... Unfortunately wasm-pack is a nightmare

Thanks for your reply and suggestions. It does give me some comfort that it was not just me who was unable to figure out the wasm-pack stuff with nix!

Regarding a self-contained "codex32 tool," agree that would be helpful.

VzxPLnHqr:
I opened a [PR to the website repo](https://github.com/apoelstra/volvelle-website/pull/1) with a flake such that `nix build .#codex32-website` will put a static copy of the website in `./results/www`.

BenWestgate:
> > ...see if I got wasm-pack working, and the answer is no... Unfortunately wasm-pack is a nightmare
> 
> Thanks for your reply and suggestions. It does give me some comfort that it was not just me who was unable to figure out the wasm-pack stuff with nix!
> 
> Regarding a self-contained "codex32 tool," agree that would be helpful.

@VzxPLnHqr: What features are you expecting a self-contained "codex32 tool" to have? I may have already made one.

VzxPLnHqr:
> What features are you expecting a self-contained "codex32 tool" to have? I may have already made one.

@BenWestgate I have not played with it too much yet, so maybe a self-hosted / static version of the website will suffice? (such as as what I added to [nixos-airgapped](https://github.com/VzxPLnHqr/nixos-airgapped) ). But a simple command line interface could be nice too. Is that what you have made already?

BenWestgate:
> > What features are you expecting a self-contained "codex32 tool" to have? I may have already made one.
> 
> @BenWestgate I have not played with it too much yet, so maybe a self-hosted / static version of the website will suffice? (such as as what I added to [nixos-airgapped](https://github.com/VzxPLnHqr/nixos-airgapped) ). But a simple command line interface could be nice too. Is that what you have made already?

Yes, https://github.com/BenWestgate/codex32/blob/master/reference/python-codex32/src/codex32.py has functions for all sorts of codex32 needs.

I started working on a Gtk GUI for importing with error correction: https://github.com/BenWestgate/bails-wallet/blob/master/bails-wallet/string_entry.py I think it was meeting most or all of the criteria of https://github.com/BlockstreamResearch/codex32/blob/master/docs/wallets.md when I last tested it.

These were originally for my project https://github.com/BenWestgate/Bails which uses the python library above to support creating codex32 backups and importing them to Bitcoin core. If andrew's bitcoin/bitcoin future PR https://github.com/bitcoin/bitcoin/pull/27351#issuecomment-2059322038 gets merged most of that code except my GUI and error correction can be deleted.

roconnor-blockstream:
`Page: 1 1` seems to now have `portraitPage begin` twice.  Presumably this is an error

apoelstra:
Oh, good catch. This appears in `title.inc.ps` but shouldn't because the default.nix code now does it.

apoelstra:
Fixed.

default.nix
```diff
@@ -0,0 +1,251 @@
...
+}:
+
+let
+  src =
+    if isNull ref
+    then lib.sourceFilesBySuffices ./. [ ".ps" ".inc" ]
+    else
+      fetchGit {
+        url = ./.;
+        inherit ref;
```
roconnor-blockstream, line 25: After reviewing https://nix.dev/manual/nix/2.22/language/builtins#builtins-fetchGit I'm pretty sure you want `rev` instead of `ref`.

apoelstra: Empirically `rev` seems to want a 40-character hash, while `ref` will accept short hashes, branch names, etc.

roconnor-blockstream:
> Oh, good catch. This appears in `title.inc.ps` but shouldn't because the default.nix code now does it.

Maybe I'm being too picky, but you patched the wrong commit.  You should have put the `title.inc.ps` patch in `add version to footer; use portraitPage and landscapePage on every page`.  However instead you patched `nix: generate postscript file from include files`.

`nix: generate postscript file from include files` was supposed to be a nop with regards to the `SSSS32.ps` file, but now that files has a diff in it, a diff that gets reversed again in `add version to footer; use portraitPage and landscapePage on every page`.

apoelstra:
I don't think you're being too picky. Though I'm surprised I made a mistake like that, given that I used `git blame` to zero in on what I thought was the correct commit, and then ran `nix-build` on every subsequent commit which showed that the SSSS32.ps file was consistent with the nixfile.

I'll fix the history. Sorry about that.

apoelstra:
Ah, I see the issue. I was misled by `git blame` (or perhaps, misled by my weird PR structure where I introduced code intending to later modify it).

Fixed.

roconnor-blockstream:
Are you familiar with `gitattributes`'s `export-subst`?

[Bitcoin uses this](https://github.com/bitcoin/bitcoin/blob/master/.gitattributes) to insert a [define `GIT_COMMIT_ID`](https://github.com/bitcoin/bitcoin/blob/master/src/clientversion.cpp#L33) into this file.  This method might make for a better way adding shortids to the (generated) postscript.

apoelstra:
No, I am not familiar -- though it looks like it needs `git-archive` to work. I am averse to using git-archive, especially the way Bitcoin Core does, because it interacts with nix in unpredictable ways. (For example in rust-bitcoinconsensus when I try to verify that the vendored Core code matches the relevant Core commit, I have a giant list of whitelisted files which seem to be .gitignored in one case but not in the other, and files which always have diffs that I can't reproduce locally, presumably because they are done by git attributes.)

I can go learn all these things if you want but I feel that my existing way, which behaves predictably and uses explicit code, will be easier to maintain.

default.nix
```diff
@@ -0,0 +1,251 @@
...
+    name = "test-single-page.ps";
+    pages = [ page ];
+  };
+in
+stdenv.mkDerivation {
+  name = "codex32${shortId}";
+
+  nativeBuildInputs = if doPdfGeneration then [ ghostscript ] else [ ];
+
+  phases = [ "buildPhase" ] ++ lib.optionals doCheck [ "checkPhase" ];
```
roconnor-blockstream, line 215: Unfortunately the `checkPhase` is never run because `doCheck` needs to be set to `true` in the `stdenv.mkDerivation` in order to run the `checkPhase`

You should change the phases line to:

    phases = [ "buildPhase" "checkPhase" ];
    inherit doCheck;

and let `doCheck` control whether or not the check phase is executed or not.

default.nix
```diff
@@ -0,0 +1,251 @@
...
+    mkdir "$out"
+    cd "$out"
+    cp "$FULL_BW" SSS32.ps
+    # Patch to include version
+    sed -i 's/(revision \(.*\))/(revision \1${shortId})/' ./SSS32.ps
+    # Produce PDF, if requested.
+    ${lib.optionalString doPdfGeneration "ps2pdf -dPDFSETTINGS=/prepress SSS32.ps"}
+  '';
+
+  checkPhase = toString
```
roconnor-blockstream, line 232: Unfortunately the check fails when it is actually enabled, at least in part because you try to run `ghostscriptTest` before it is defined.

default.nix
```diff
@@ -0,0 +1,251 @@
...
+    (map
+      (page: "ghostscriptTest ${checkSinglePage page}")
```
roconnor-blockstream, line 234: There is another issue with this design.  Because `checkPhase` contains references to the single pages, those single pages must be built, whether or not they will be checked.  In this case I suggest taking the unusual step of conditionally excluding the entire contents of the `checkPhase` value if `doCheck` is false.

Alternatively you could replace `fullBooklet.pages` with `[]` when `doCheck` is false.

roconnor-blockstream (changes requested):

default.nix
```diff
@@ -0,0 +1,251 @@
...
+  name = "codex32${shortId}";
+
+  nativeBuildInputs = if doPdfGeneration then [ ghostscript ] else [ ];
+
+  phases = [ "buildPhase" ] ++ lib.optionals doCheck [ "checkPhase" ];
+
+  buildPhase = ''
+    set -e
+
+    FULL_BW="${renderBooklet fullBooklet}"
```
roconnor-blockstream, line 220: The way nix derivations work is that, more or less, every value in the derivation becomes an environment variable.

Therefore, it would make more sense for you to define `FULL_BW` as an entry in the derivation rather that setting it in the build phase.

PR #70, opened 2025-04-04
Haskell implementation of Codex32

roconnor-blockstream:
Includes an executable program `codex32` whose command `correct` implements the codex32 error correction algorithm.

### DIFF
--- a/haskell/exec/Main.hs
+++ b/haskell/exec/Main.hs
@@ -1,0 +1,119 @@
+module Main where
+
+import Control.Applicative ((<**>), (<|>), many)
+import Control.Arrow ((+++))
+import Control.Monad (unless)
+import Data.List (find,intercalate, nub)
+import Data.Monoid ((<>))
+import Data.Text (pack)
+import qualified Options.Applicative as Opt
+import qualified Options.Applicative.Types as Opt
+import qualified Prettyprinter as PP
+import qualified Prettyprinter.Util as PP
+import qualified System.Exit as Sys
+
+import Codex32
+import Codex32.Error
+import Codex32.Word5
+import Codex32.Polynomial
+
+data AdvancedOptions = AdvancedOptions { optLength :: Int
+                                       , optOverrideLength :: Bool
+                                       , optErasures :: [Int]
+                                       , optSpec :: Spec
+                                       , optResidue :: Poly Word5
+                                       }
+
+data Command = Simple String
+             | Advanced AdvancedOptions
+
+residueReader :: Opt.ReadM (Spec, [Word5])
+residueReader = do
+  result <- Opt.eitherReader parse
+  let len = length result
+  spec <- maybe (failLength len) return $ find (\spec -> len == specDegree spec) specs
+  return (spec, result)
+ where
+  parse = (errMsg +++ reverse) . fromString
+  errMsg c = "Illegal Bech32 character " ++ show c ++ "."
+  failLength len = fail $ "Residue length must be " ++ intercalate " or " (show <$> validDegrees) ++ "."
+  validDegrees = specDegree <$> specs
+  specs = [codex32Spec, codex32LongSpec]
+
+codex32SimpleParser :: Opt.Parser Command
+codex32SimpleParser = Simple <$> Opt.strArgument (Opt.metavar "CODEX32_STRING" <> Opt.help "Codex32 string to correct")
+
+codex32AdvancedParser :: Opt.Parser Command
+codex32AdvancedParser = Advanced <$>
+  (mkAdvancedOptions <$> Opt.option Opt.auto (Opt.long "len" <> Opt.metavar metaLength <> Opt.help "Total length of codex32 string")
+                     <*> Opt.switch (Opt.long overrideOptName <> Opt.hidden)
+                     <*> many (Opt.option Opt.auto (Opt.short 'e' <> Opt.metavar "ERASURE_LOCATION" <> Opt.help "Location of unreadable character (can be repeated)"))
+                     <*> Opt.argument residueReader (Opt.metavar metaResidue <> Opt.help "Residue from worksheet"))
+ where
+  mkAdvancedOptions l ol e (s, r) = AdvancedOptions l ol e s r
+
+metaLength = "LENGTH"
+overrideOptName = "override_length"
+metaResidue = "RESIDUE"
+
+codex32CorrectParser :: Opt.ParserInfo Command
+codex32CorrectParser = Opt.info ((codex32SimpleParser <|> codex32AdvancedParser) <**> Opt.helper) (Opt.progDescDoc desc)
+ where
+  pretty = PP.reflow . pack
+  desc = Just $ pretty "Error correct codex32 strings."
+             <> PP.line <> PP.line
+             <> pretty "Pass an 'ms1' string on the command line to try and output an error-corrected version of that string."
+             <> PP.line <> PP.line
+             <> pretty "Alternatively, if you are verifying your checksum using the worksheet from the Codex32 booklet, you can pass '--len 48' followed by the 13 character residue you computed on your worksheet."
+             <> PP.space <> PP.space
+             <> pretty "If you have unreadable characters in your string, use your best guess and add '-e <location>' to let the tool know which box contained your unreadable character."
+             <> PP.space <> PP.space
+             <> pretty "If error correction is possible, the program will output instructions on which character locations contain errors and what values to add, using the booklet's addition wheel, in order to correct the error."
+             <> PP.space <> PP.space
+             <> pretty "With this method, the computer can help you correct your codex32 string without even needing to know what the string is."
+
+codex32CorrectOptions :: Opt.Parser Command
+codex32CorrectOptions = Opt.subparser (Opt.command "correct" codex32CorrectParser)
+
+codex32Options :: Opt.ParserInfo Command
+codex32Options = Opt.info (codex32CorrectOptions <**> Opt.helper) mempty
+
+codex32Prefs :: Opt.ParserPrefs
+codex32Prefs = Opt.prefs Opt.showHelpOnEmpty
+
+formatCorrections :: Int -> [(Int, Word5)] -> String
+formatCorrections _ [] = "No errors found.  Residue is correct."
+formatCorrections len corrections = unlines (header : (fmt <$> corrections))
+ where
+  header = show len ++ " errors found.  Make the following corrections."
+  fmt (ix, delta) = "Add " ++ show (toChar delta) ++ " to position " ++ show (len - ix) ++ "."
+
+main :: IO ()
+main = Opt.customExecParser codex32Prefs codex32Options >>= run
+
+run :: Command -> IO a
+run (Simple codex32Str) =
+  case (correctCodex32String codex32Str) of
+    Nothing -> putStrLn "Failed to error correct string" >> Sys.exitFailure
+    Just str -> putStrLn str >> Sys.exitSuccess
+run (Advanced options) | 13 < length erasureIxs = failWith "No more than 13 -e options are allowed."
+                       | not (optOverrideLength options) && bitsize `notElem` [128, 256, 512] = failWith $ "Unusual bitsize found.  Override with --" ++ overrideOptName ++ "."
+                       | 5 <= padding = failWith $ "Invalid " ++ metaLength ++ "."
+                       | len < 48 = failWith $ metaLength ++ " too short."
+                       | 127 < len = failWith $ metaLength ++ " too long."
+                       | specDataLength spec < 6 + payloadLength = failWith $ metaLength ++ " too long for " ++ show (length residue) ++ " character " ++ metaResidue ++ "."
+                       | 15 == length residue && len < 99 = failWith $ metaLength ++ " too short for " ++ show (length residue) ++ " character " ++ metaResidue ++ "."
+                       | otherwise = format result
+ where
+  erasureIxs = nub (optErasures options)
+  residue = optResidue options
+  len = optLength options
+  spec = optSpec options
+  dataLength = len - length (specPrefix spec) - 1
+  payloadLength = dataLength - 6 - length (specTarget spec)
+  (bytesize, padding) = (payloadLength * 5) `divMod` 8
+  bitsize = bytesize * 8
+  failWith str = Opt.handleParseResult . Opt.Failure $ Opt.parserFailure codex32Prefs codex32Options (Opt.ErrorMsg str) [Opt.Context "correct" codex32CorrectParser]
+  result = errorCorrections (optSpec options) erasureIxs residue
+  format Nothing = putStrLn "Too many errors.  Unable to correct." >> Sys.exitFailure
+  format (Just corrections) = putStr (formatCorrections len corrections) >> Sys.exitSuccess
--- a/haskell/exec/Main.hs
+++ b/haskell/exec/Main.hs
@@ -20,37 +20,27 @@ import Codex32.Polynomial
 data AdvancedOptions = AdvancedOptions { optLength :: Int
                                        , optOverrideLength :: Bool
                                        , optErasures :: [Int]
-                                       , optSpec :: Spec
                                        , optResidue :: Poly Word5
                                        }
 
 data Command = Simple String
              | Advanced AdvancedOptions
 
-residueReader :: Opt.ReadM (Spec, [Word5])
-residueReader = do
-  result <- Opt.eitherReader parse
-  let len = length result
-  spec <- maybe (failLength len) return $ find (\spec -> len == specDegree spec) specs
-  return (spec, result)
+residueReader :: Opt.ReadM [Word5]
+residueReader = Opt.eitherReader parse
  where
   parse = (errMsg +++ reverse) . fromString
   errMsg c = "Illegal Bech32 character " ++ show c ++ "."
-  failLength len = fail $ "Residue length must be " ++ intercalate " or " (show <$> validDegrees) ++ "."
-  validDegrees = specDegree <$> specs
-  specs = [codex32Spec, codex32LongSpec]
 
 codex32SimpleParser :: Opt.Parser Command
 codex32SimpleParser = Simple <$> Opt.strArgument (Opt.metavar "CODEX32_STRING" <> Opt.help "Codex32 string to correct")
 
 codex32AdvancedParser :: Opt.Parser Command
 codex32AdvancedParser = Advanced <$>
-  (mkAdvancedOptions <$> Opt.option Opt.auto (Opt.long "len" <> Opt.metavar metaLength <> Opt.help "Total length of codex32 string")
-                     <*> Opt.switch (Opt.long overrideOptName <> Opt.hidden)
-                     <*> many (Opt.option Opt.auto (Opt.short 'e' <> Opt.metavar "ERASURE_LOCATION" <> Opt.help "Location of unreadable character (can be repeated)"))
-                     <*> Opt.argument residueReader (Opt.metavar metaResidue <> Opt.help "Residue from worksheet"))
- where
-  mkAdvancedOptions l ol e (s, r) = AdvancedOptions l ol e s r
+  (AdvancedOptions <$> Opt.option Opt.auto (Opt.long "len" <> Opt.metavar metaLength <> Opt.help "Total length of codex32 string")
+                   <*> Opt.switch (Opt.long overrideOptName <> Opt.hidden)
+                   <*> many (Opt.option Opt.auto (Opt.short 'e' <> Opt.metavar "ERASURE_LOCATION" <> Opt.help "Location of unreadable character (can be repeated)"))
+                   <*> Opt.argument residueReader (Opt.metavar metaResidue <> Opt.help "Residue from worksheet"))
 
 metaLength = "LENGTH"
 overrideOptName = "override_length"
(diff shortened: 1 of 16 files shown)


roconnor-blockstream:
I've added a bit more to the help documentation.

haskell/exec/Main.hs
```diff
@@ -0,0 +1,119 @@
...
+
+main :: IO ()
+main = Opt.customExecParser codex32Prefs codex32Options >>= run
+
+run :: Command -> IO a
+run (Simple codex32Str) =
+  case (correctCodex32String codex32Str) of
+    Nothing -> putStrLn "Failed to error correct string" >> Sys.exitFailure
+    Just str -> putStrLn str >> Sys.exitSuccess
+run (Advanced options) | 13 < length erasureIxs = failWith "No more than 13 -e options are allowed."
```
BenWestgate: This shouldn't be a constant, the long checksum can correct 15 consecutive erasures.

haskell/exec/Main.hs
```diff
@@ -0,0 +1,119 @@
...
+                       | not (optOverrideLength options) && bitsize `notElem` [128, 256, 512] = failWith $ "Unusual bitsize found.  Override with --" ++ overrideOptName ++ "."
+                       | 5 <= padding = failWith $ "Invalid " ++ metaLength ++ "."
+                       | len < 48 = failWith $ metaLength ++ " too short."
+                       | 127 < len = failWith $ metaLength ++ " too long."
+                       | specDataLength spec < 6 + payloadLength = failWith $ metaLength ++ " too long for " ++ show (length residue) ++ " character " ++ metaResidue ++ "."
+                       | 15 == length residue && len < 99 = failWith $ metaLength ++ " too short for " ++ show (length residue) ++ " character " ++ metaResidue ++ "."
```
BenWestgate: Don't you need the equivalent 13 == length residue && len > 74 (i think)...
There's a no man's land length between the two checksum types where the data isn't valid. (A good reason to change BIP93 to support fewer lengths, perhaps the BIP39 ones and 512.)

haskell/exec/Main.hs
```diff
@@ -0,0 +1,119 @@
...
+  residue = optResidue options
+  len = optLength options
+  spec = optSpec options
+  dataLength = len - length (specPrefix spec) - 1
+  payloadLength = dataLength - 6 - length (specTarget spec)
+  (bytesize, padding) = (payloadLength * 5) `divMod` 8
+  bitsize = bytesize * 8
+  failWith str = Opt.handleParseResult . Opt.Failure $ Opt.parserFailure codex32Prefs codex32Options (Opt.ErrorMsg str) [Opt.Context "correct" codex32CorrectParser]
+  result = errorCorrections (optSpec options) erasureIxs residue
+  format Nothing = putStrLn "Too many errors.  Unable to correct." >> Sys.exitFailure
```
BenWestgate, line 112: This is a bit pessimistic, if erasure locations were provided the user assuming some likely values could put a correction in reach. "Too many errors and erasures." Reminds the user to try specifying less erasures if possible.

haskell/test/Tests.hs
```diff
@@ -0,0 +1,73 @@
...
+correctableBCHErrorVector spec length = do
+  (errorCount, erasureCount) <- elements distribution
+  let remainderCount = length - errorCount - erasureCount
+  errorString <- replicateM errorCount genError
+  let erasureString = replicate erasureCount Nothing
+  let remainderString = replicate remainderCount 0
+  shuffle $ erasureString ++ fmap Just (errorString ++ remainderString)
+ where
+  distance = specDistance spec
+  distribution = do
```
BenWestgate, line 37: Since BIP93 explicitly mentioned contiguous erasure correction perhaps it would be good to test that specific pattern up to the checksum length?  I'm not sure how it interacts with simultaneous error correction limits but surely better per erasure than randomly spaced.

BenWestgate (commented):
I reviewed the tests, main, codex32.hs and error.hs.

I'll try to use this to improve my brute force insert and delete correction tool this year.

I built it I wasn't able to pass erasures with `-e 5` on the command line. I'm probably doing the syntax wrong.

A better syntax, imho, I used in my brute force tool was to assume '?' in the codex32 string passed were erasures. Seems clunky to specify erasure locations individually.

roconnor-blockstream:
> I built it I wasn't able to pass erasures with -e 5 on the command line. I'm probably doing the syntax wrong.

> Usage: codex32 correct (CODEX32_STRING | --len LENGTH [-e ERASURE_LOCATION]  RESIDUE)

You have to pass the `--len` argument before `-e` arguments.  e.g. `--len 48` for 128 bit secrets.

Maybe I should add a choice between specifying the string length or the number of bits in the secret.

Edit: Most definitely the help text needs to illustrate the two different modes of operation and note that you can use ? to mark erasures in the simple mode.

haskell/test/Tests.hs
roconnor-blockstream: Below `correctableLinearErrorVector` creates a test template where a consecutive series of `degree` many erasures is added somewhere in the middle of the string.

roconnor-blockstream:
I converted it to draft because I need to squash my changes, and I want to fix up a few more things.  However it is still reviewable.

haskell/exec/Main.hs
roconnor-blockstream: Fixed.

haskell/exec/Main.hs
roconnor-blockstream: With fresh eyes, I've rearchitected how the advanced command line arguments are transformed into the specification data needed for error correction.

haskell/exec/Main.hs
roconnor-blockstream: I still need to address this.

roconnor-blockstream:
> A better syntax, imho, I used in my brute force tool was to assume '?' in the codex32 string passed were erasures. Seems clunky to specify erasure locations individually.

In the the simple error correction mode, you just use '?' for erasures like you imagine.

The advanced mode is where you are doing paper computing and you got an checksum error, and you don't want to tell your computer what your share is.  With the advanced mode you only tell the computer what you incorrect 13 (or 15!) character residue is that you computed was, then you tell the computer where you think erasures might be, then the computer blindly tells you how to correct your string without ever needing access to your share data.

The advanced mode is a way of doing error correction computations while minimizing the information input into a computer.  Under the assumption that all errors are equally likely the computer learns no information about your secret share.  However, that assumption probably isn't true and in practice tiny fractions of bits of information might in theory be inferred by a malicious computer about the characters at the locations where errors were found.

haskell/exec/Main.hs
BenWestgate: Pessimistic because our [wallets.md](https://github.com/BlockstreamResearch/codex32/blob/master/docs/wallets.md) tells developers to automatically replace invalid characters with '?' in many cases: non-numeric threshold, non-bech32 character, non-"s" index if k=0, repeated share indices (after user affirms they aren't repeating an already entered share).

Any of these silently placed '?' could probably be filled by more carefully reading or typing or cross referencing another share. Additionally, if the identifier was the default bip32 master fingerprint, those erasures are easily filled out of band.

I suggest:

"Too many errors and erasures. Replace some erasures with your best guesses and try again."

BenWestgate:
> > I built it I wasn't able to pass erasures with -e 5 on the command line. I'm probably doing the syntax wrong.
> 
> > Usage: codex32 correct (CODEX32_STRING | --len LENGTH [-e ERASURE_LOCATION]  RESIDUE)
> 
> You have to pass the `--len` argument before `-e` arguments. e.g. `--len 48` for 128 bit secrets.
> 
> Maybe I should add a choice between specifying the string length or the number of bits in the secret.
> 
> Edit: Most definitely the help text needs to illustrate the two different modes of operation and note that you can use ? to mark erasures in the simple mode.

in `wallets.md` we wrote:
> ECWs MAY assume the correct length is the closest of 48 or 74.

I've since I've updated guidance to say generating other lengths than 128, 256, 512 is "NOT RECOMMENDED", as well as supporting their import is "NOT RECOMMENDED", then we can strengthen this to ECWs SHOULD assume the correct length is the closest of 48, 74 or (whatever length 512 is).

MAY support for weird lengths like 17 bytes or 31 byte seeds, SHOULD import them if the checksum passes, but MAY error correct them first as if they were 16 or 32 byte seeds. Essentially, no error correction at all, assuming they're mistranscriptions of 16 and 32 byte seeds respectively, only testing the original length (and others nearby) if there are no valid corrections within edit distance limits for 16 and 32 byte candidates.

This was originally in regards to insert/delete correction, but it seems it applies just as well to erasure and error correction in your implementation. Fixing a delete is just a matter of trying 48 erasure positions.

BenWestgate:
> You have to pass the `--len` argument before `-e` arguments. e.g. `--len 48` for 128 bit secrets.
> 
> Maybe I should add a choice between specifying the string length or the number of bits in the secret.

Eliminate the string length parameter entirely.

It is specialist knowledge, I barely remember the length of 32- or 64-byte codex32 strings off the top of my head and wrote a [`codex32`](https://pypi.org/project/codex32/) PyPI package and years ago a codex32 error correcting wallet. We shouldn't expect users to know. `wallets.md` says "error correcting wallets MAY assume the correct length is the closest of..." So do this, otherwise let them specify a seed length in bytes, only whole bytes are valid.

No string length parameter as there are two invalid lengths in the middle. Prevent user mistakes. *if they have to count the characters, they definitely don't know their seed byte length, so we should assume for them.

BenWestgate:
> The advanced mode is a way of doing error correction computations while minimizing the information input into a computer.

I will probably need to use it this way in my wallet because it runs on a restricted platform [Tails OS](https://tails.net/) and users can't by default install the packages needed to build Haskell code. Would ship a signed binary, and then not hand it secret data during error correction.

> the computer blindly tells you how to correct your string without ever needing access to your share data.

It's also possible to one-time-pad encrypt the string by a random valid string and pass that to simple mode. Not sure what is easier by hand. Assume the encryption string is generated with a different offline PC. Do you know? That may affect whether "advanced mode" is useful or not.

roconnor-blockstream:
> Eliminate the string length parameter entirely.

I think you are misunderstanding the operation of the correction tool:

> Usage: codex32 correct (CODEX32_STRING | --len LENGTH [-e ERASURE_LOCATION] RESIDUE)

The `|` is "or".  Here are some example uses (which I ought to add to the documentation).

    $ codex32 correct ms12namea320zyxwvut5rqpnmlkjhgfedcaxrpp870hkkqrm
    ms12namea320zyxwvutsrqpnmlkjhgfedcaxrpp870hkkqrm

The above is the simple usage where you just provide a string.  Notice I've made an error: in the middle of the string I have written a '5' when it was supposed to be an 'S', a common error.  The corrected string is printed.  No lengths are specified.

However, if I wanted to use the "advanced" mode, I would use the worksheet from the booklet to compute a residue of "C0T9WYQDDVRPP".  The residue is supposed to be "SECRETSHARE32", so clearly I have made some sort of mistake.

If I had a proper wallet, I would just pass in my erroneous string to, like in the simple case, since I'm going to trust my wallet with my secret anyways.  Also, for such a trivial mistake, I can look up this residue in @apoelstra's table of simple error corrections, but let's continue anyways.

But let's say, for some reason, I don't have an error correcting wallet and I don't have @apoelstra's table (or I didn't find my residue in his tables).  I can still correct my error using the computer while leaking minimal information using the "advanced" mode and my paper computed residue.

    $ codex32 correct --len 48 C0T9WYQDDVRPP
    1 errors found.  Make the following corrections.
    Add 'y' to position 20.

I know my string is 48 characters long, because I just used the worksheet in which the character positions are numbered.  The "advanced" mode tells me that to fix the error, I have to add 'Y' to the character at position 20.  Looking up box 20 on my worksheet I used, I see position 20 is the location of '5'.  I use the addition table or addition wheel to add 'Y' to '5' and I get 'S'.  Substituting 'S' for '5' gets me the corrected string.

In order to compute the position of the error we need to know the length of the original codex32 string.  Because the advanced mode only receives the residue, it has no way of knowing how long the original string was.  (Technically I can tell you where the error is by counting backwards from the end of the string without knowing the string's length, and maybe I should add this mode as well.)

How much information did the computer learn in this case?  Well the computer learned that I made an error in position 20, and the correction was to add 'Y' to whatever character was there.  That means the computer knows that I made and error such as:

* 'A' was replaced with 'E' or vice versa.
* 'C' was replaced with 'U' or vice versa.
* 'D' was replaced with 'F' or vice versa.
* ...
* 'S' was replaced with '5' or vice versa.
* ...

Presumably if you do a symbol likedness comparison for all these pairs, you will find that '5' and 'S' are most similar.  Therefore the computer will learn that your share's character at position 20 is somewhat likely to be a '5' or an 'S'.  That is at most 2.5 bits of information.  It's going to be less than that because the computer isn't 100% sure it is a '5' or an 'S'.  Maybe the error was swapping and 'H' with and 'N' by writing a sloppy cross bar.

Giving the computer access to those bits of information isn't great, but it is certainly better than the simple mode where the computer learns the entire share outright.  Though again, if you are using a wallet you are trusting with your whole secret anyways, there is no problem using the "simple" mode to do error correction.

Since we will expect wallets to do error correction, I'm having trouble imagining where advanced mode would be useful in practice.  But it is nice to see that advanced mode does work.

Edit: I forgot to add: blinding the share doesn't help because even in that case the computer will learn that I made an error in position 20 whose correction was to add 'Y' to whatever character was there. Since blinding is strictly more work, it is preferable to just to pass the computed residue to the command line using the advanced mode.

BenWestgate:
> Notice I've made an error: The corrected string is printed. No lengths are specified.

Off-topic question: I'm writing an ECW that may use your tool.  In the case of 5 substitutions when the "correction" is wrong, do 9 symbols always differ from the correct string? And what is the distribution of those differences? As in, for what distance is our code HD=10?  (able to detect 9 errors, so the correction won't have miscorrections clustered tighter than that).

Knowing determines how coarsely UIs should highlight the general location of corrections. I am concerned highlighting only corrected characters misleads the user as the whole corrected string must match the original.

If you don't know, just say, and I'll use your tool to run a simulation of 5+ errors.

roconnor-blockstream:
I expect 9 symbols to usually differ, but I don't know.

BenWestgate:
> I know my string is 48 characters long, because I just used the worksheet in which the character positions are numbered.

In this case, specifying the length is ok if the only way to get a residue makes it clear. But if you can say 28th from the right, then there's no possibility of the user typing the wrong length and easier to script calls to it.

> I expect 9 symbols to usually differ, but I don't know.

If less than 9 differed, we would expect the error to be detected, so I think it's by definition always 9 errors in a wrong correction, as it can only correct 4 characters. And then the clustering of them will always be more diffuse than the distance our code is HD=10 for.  Which needs some simulation to learn.

But say an error exists every 5 characters, your tool can give 4 error locations (diff between input and correction), **all wrong characters**, spaced on average 11.25 characters apart. To highlight all errors if evenly spaced, highlighting a three 4-character window (12 symbols) that contains a correction might work.

But since for small numbers of errors, the "corrections" are always wrong, it seems smart to highlight those in bright red and the rest of the group in darker red.

BenWestgate:
> Therefore the computer will learn that your share's character at position 20...

All the more reason to not tell the computer a length and say "position at len - 28" let the user subtract or count the position from right to left, it's not hard. In fact, we could argue the book numbered the characters backwards if residues predict error locations in relation to the end of the string but not from the beginning. So PR the booklet and drop the --len parameter entirely may be best.

> I'm having trouble imagining where advanced mode would be useful in practice. But it is nice to see that advanced mode does work.

All I can think of is the user lost some tiles from a metal backup or water damaged written ones.  This is why we should say the erasure limit is up to 13 or 15, as burst erasures will be most common. They don't need to spend so they don't need to reveal the secret to new hardware so they use advanced mode to restore share integrity.

It would be hard to know there were errors needing correction, without annual check ups. And we should recommend to KEEP BOTH copies in case the correction is wrong.  But if the user suspects it is only erasures and enters what remains extra carefully (errors in the data may yield false erasure corrections.) They can achieve their objective. Fairly niche but it might occur if users actually do integrity checks more often than they recover their seed.

roconnor-blockstream:
> (how would you even know there were substitution errors??)

I just showed you an example:

```
$ codex32 correct --len 48 C0T9WYQDDVRPP
1 errors found.  Make the following corrections.
Add 'y' to position 20.
```

This corrects a substitution error in the string `ms12namea320zyxwvut5rqpnmlkjhgfedcaxrpp870hkkqrm`.

BenWestgate:
> > (how would you even know there were substitution errors??)

I meant _how would_ the user know by looking their string needed corrections in the first place. But now, I realize if the checksum worksheet is performed as part of an integrity checkup it has the wrong residue which indicates using advanced mode to correct it.

>...without annual check ups. And we should recommend to KEEP BOTH copies in case the correction is wrong. But if the user suspects it is only erasures and enters what remains extra carefully (errors in the data may yield false erasure corrections.) They can achieve their objective. Fairly niche but it might occur if users actually do integrity checks more often than they recover their seed.

I'm uncomfortable with this use-case. If I find errors or erasures, and use math to correct them (as opposed to finding a backup copy) I would want to recover my wallet to prove to myself the corrections are the right ones.  Errors in one backup correlate with errors in others (in my experience recovering multi-sig wallets for people) so I wouldn't want to let the data sit and possibly accumulate more errors or erasures with time until it's uncorrectable.

An exception would be a codex32 secret with:
1. BIP32 fingerprint as the identifier
2. CRC code as padding bits
If the identifier matches what my watch-only has for key origin info, and the CRC passes, I would feel confident enough to let it sit.  This is about 27-33 extra bits of assurance than the codex32 checksum alone.

roconnor-blockstream:
> I meant how would the user know by looking their string needed corrections in the first place. But now, I realize if the checksum worksheet is performed as part of an integrity checkup it has the wrong residue which indicates using advanced mode to correct it.

Yeah, the user would run a quickcheck worksheet with a 2 character checksum.  If that fails, they'd have to do the full worksheet to compute the full residue.

> Errors in one backup correlate with errors in others (in my experience recovering multi-sig wallets for people)

Can you say more about this?

BenWestgate:
> > Errors in one backup correlate with errors in others (in my experience recovering multi-sig wallets for people)
> 
> Can you say more about this?

Sure.  I helped someone recover a [Yeticold](https://github.com/JWWeatherman/yeticold/blob/master/templates/HotSeedTemplate.txt) wallet.  It is a 3-of-7 multisig, with a rather unsophisticated seed backup of WIF format expanded to NATO phonetic words.  Every 4 words a check word (sum mod 58 lol).  Terrible format really, was rejected as a BIP, but that's we had to recover.

I needed to read 6 of the 7 seeds before I had three WIF seeds Bitcoin Core accepted.  Similar problems with each of the 3 invalid seeds: illegible/missing/extra/transposed symbols.

It's what inspired me to want to make a CodexQR #66 format for converting codex32 shares to compact QR codes that can be hand drawn in a few minutes, some users have terrible penmanship and it can lead to loss of bitcoin, especially for heirs.

Operator fatigue contributed to this failure. It's hard to write 455 random words, most in ALL CAPS, without breaks.

Further, the confirmation UI must have allowed him to fix "typos" and errors in his on-screen entry without scaring him enough to also correct his paper copy.

BenWestgate:
Is this ready for me to test and review again?

> I converted it to draft because I need to squash my changes, and I want to fix up a few more things. However it is still reviewable.

@roconnor-blockstream Is this ready for me to test again?  If I add code for insert/delete corrections we might have better data on whether https://github.com/bitcoin/bips/pull/2077 is the right approach.

I'm also considering what you said in https://github.com/bitcoin/bips/pull/2040 and making which checksum to use be an application specific choice rather than specification level.

As that also avoids the "ms" exception, every `hrp` can define it's own checksum rules.  After all, without the HRP, we don't even know a string is codex32 or Bech32 and it could even validate for both.

If this is the best approach then there's no need to restrict "ms" lengths unless it helps error correction significantly.

BenWestgate:
I tested bounded insertion/deletion recovery around this PR’s fixed BCH corrector using [your proposed](https://github.com/bitcoin/bips/pull/2258#issuecomment-5411804501) restricted `ms` sizes 48, 54, 61, 67, 74, and 127.
Using one cumulative false-correction bound `p <= 1e-5` for all searched lengths, I found unknown length recovery can search:
- 48, 74, and 127: I + O <= 4
- 54, 61, and 67: I + O <= 3
- all six lengths: GI + GO <= 2 for four-character group indels

The worst cumulative bound was 8.94e-6, below the 1e-5 target. All cases where a 48-character reconstruction was possible completed within 9 seconds on a Ryzen 7 7735U. Giving every valid length I + O <= 4 remained probability-safe but exceeded 18 seconds in some cases.

Consequently, the [wallets.md](https://github.com/BlockstreamResearch/codex32/blob/master/docs/wallets.md) advice that correction may assume the closest of 48 or 74 should be updated as searching 54, 61and 67 at one less character indel adds negligible time or false-corrections.

A known correct length does not help search beyond I + O <= 4: known length I + O <= 5 exceeds 10 seconds and requires `p <= 1e-4`.

A known correct header, cuts known length search times by ~1/3, but I did not attempt to search subsequent shares deeper.

All the character indel and four-character group indel search spaces are in addition to fixed length BCH corrections subject to `p <= 1e-5` for example 2GO has no BCH capacity left, while 2I+2O can also correct one substitution but two exceeds false-correction probability.

PR #71, opened 2025-10-17
doc: Add unshared secret generation to wallets.md

### LINKED ISSUE #64: Suggest standard implementation for electronic share set generation
**Summarizes Issue #57:**

While [codex32](https://github.com/bitcoin/bips/blob/master/bip-0093.mediawiki) is designed to be computed by hand, should it become popular, it's likely most backups will be made with electronics. This offers opportunities to increase error detection and presents liabilities from trusting the randomness of a device. This proposal seeks to maximize benefits of electronics while reducing their downside by being deterministic and easily [auditable](https://github.com/oleganza/bitcoin-papers/blob/master/AuditableBitcoinWallets.md) so the device's entropy is not a single point of failure.

**Generating Shares**
*For a fresh master seed*
**Inputs:**

- **App entropy** an extended private masterkey generated by the wallet. It must be displayed unconditionally so the wallet does not know if it will be audited.
- **User entropy** a string provided by the user that allows improving security in case app entropy generator has a backdoor. This could be a one-time text entered by the user only for this purpose, or a passphrase containing enough entropy that is used later for other purposes.
- **k** the threshold to recover the codex32 secret. 
- **n** the total number of shares in the backup.

*Optional inputs*:
- **Identifier** sets a specific 4 character bech32 identifier for the backup, overriding the deterministic one.
- **Seed length** sets a specific master seed byte length, overriding the default 16-bytes.
- **Fixed share indices** forces up to n of the indices to be the ones provided.
- **Existing codex32 strings** forces up to k of the shares to be the ones provided. Overrides seed length, identifier and k.

**Outputs:**

- **Fresh master seed**, the BIP32 master seed of the backup.
- **New shares**, n newly generated shares with shuffled indices. (excludes existing)

*For an existing master seed*
     - Inputs and outputs are the same except:
**Inputs:**
- **Existing codex32 strings**: one must be the *codex32 secret* or else a valid threshold length set of shares must be given.
- **App entropy** is the private masterkey of the existing master seed.
- **User entropy** becomes optional unless *re*-sharing a master seed with an existing codex32 share backup. Then it must be a "unique string". The app may help ensure uniqueness with a monotonic counter or date which must be unconditionally displayed.

**PROPOSAL**

- **Share indices**

In order to prevent the order of indices used from leaking secret information, the available indices (after excluding 's' and any existing shares passed) are deterministically shuffled as follows:

```python3
def shuffle_indices(index_seed, indices):
    from cryptography.hazmat.primitives.ciphers import Cipher, algorithms
    """
    Shuffle indices deterministically using provided key with ChaCha20.

    :param index_seed: The ChaCha20 key for deterministic shuffling.
    :param indices: Characters to be shuffled as a string or list.
    :return: List of shuffled characters sorted by assigned values.
    """
    algorithm = algorithms.ChaCha20(index_seed, bytes(16))
    keystream = Cipher(algorithm, mode=None).encryptor()
    counter = 0  # Counter to track current position in the keystream.
    value = b""  # Storage for the assigned random byte.
    block = b""  # Holds the latest keystream block.
    assigned_values = {}  # Dictionary to store chars and their values.
    for char in indices:
        # Ensure new random value is generated if there is a collision.
        while value in assigned_values.values() or not value:
            if not counter % 64:  # Get new 64-byte block per 64 count.
                block = keystream.update(bytes(64))  # ChaCha20 block.
            value = block[counter % 64:counter % 64 + 1]  # Rand byte.
            counter += 1
        assigned_values[char] = value
    return sorted(assigned_values.keys(), key=lambda x: assigned_values[x])
```

- **Padding bits**

In order to prevent padding bits from leaking 2-4 bits per generated share, these must also be set deterministically. While all zeros works, it's wasteful as some error detection is possible with an ECC code. (potentially bit error correction for high k values.) This ECC padding is guaranteed to detect up to 1 bit error or 2 sequential bit errors:

```python3
def ecc_padding(data):
    """
    Calculate and return a byte with concatenated parity bits.

    :param data: Bytes of seed_len, hrp, k, ident, index and payload.
    :return: Byte with concatenated parity in most significant bits.
    """
    # Count mod 2 the number of set (1) bits in the byte data
    parity = bin(int.from_bytes(data)).count('1') % 2 << 7
    # Count mod 2 the number of set (1) even bits in the byte data
    parity += bin(int.from_bytes(data))[::2].count('1') % 2 << 6
    # Count mod 2 the number of set (1) odd bits in the byte data
    parity += bin(int.from_bytes(data))[3::2].count('1') % 2 << 5
    # Count mod 2 the number of set (1) third bits in the byte data
    parity += bin(int.from_bytes(data))[::3].count('1') % 2 << 4

    return parity.to_bytes()
```

This may help if a share is damaged beyond the correction ability of the codex32 checksum by reducing correction candidates by 1/4 to 1/16. It also checksums string length that BCH checksums can't offer error detection guarantees about.

- **Identifier**

There are two good ideas for setting the default ID so that 1) users are not burdened with creating and entering it, 2) 20-bits of the secret aren't leaked, and 3) it's easy to identify which wallet or descriptor the shares belong to or vice versa.

1. **BIP32 fingerprint in bech32**

  - Allows matching the first 5 hex characters from descriptor key origin information to the codex32 share identifier to confirm they correspond.
  - Allows error correcting these 4 characters by converting a known corresponding fingerprint. Further helping emergency recovery.
  - As a cryptographic checksum, it detects malicious tampering of shares or invalid combinations as the derived seed's fingerprint won't match.

2. **Encrypted BIP32 fingerprint in bech32**
 - Allows selectively revealing the correspondence between shares and descriptors only to those with the `unique_string`.
 - Allows creating _unique_ identifiers for re-sharing that still offer the benefits of fingerprint above if `unique_string` is remembered or stored.
 - Includes `k` and `len(payload)` in KDF's `salt` allowing detection of errors in these with known corresponding BIP32 fingerprint and `unique_string`.

```python3
def encrypt_fingerprint(master_seed, k, unique_string=""):
    """
    Encrypt the MS32 fingerprint using a unique string and header data.

    :param master_seed: The master seed used for BIP32 fingerprint.
    :param k: The threshold parameter as a string.
    :param unique_string: Optional unique string encryption password.
    :return: Encrypted fingerprint as a bech32 string.
    """
    password = bytes(unique_string, "utf")
    salt = len(master_seed).to_bytes(1, "big") + bytes(k, "utf")
    enc_key = convertbits(hashlib.scrypt(
        password, salt=salt, n=2 ** 20, r=8, p=1, maxmem=1025 ** 3, dklen=3),
        8, 5, pad=False)
    new_id = [x ^ y for x, y in zip(ms32_fingerprint(master_seed), enc_key)]
    return "".join([CHARSET[d] for d in new_id])

```
I prefer using the encrypted fingerprint everywhere for more privacy and its ability to detect errors in k and string length which _may_ help emergency recovery. It's also slower to compute due to KDF so harder to grind data into.

- **Random share payloads**

`k` shares are generated (less any provided existing codex32 strings). Generating random payloads is the largest potential leak. My proposal has two steps:

1. User entropy and an extended private masterkey (_xprv_) from app entropy (or an existing master seed) are fed to a computationally expensive KDF to produce a 128 byte `derived_key`. 
2. Truncated HMAC-SHA512 is used with `key=derived_key` and a unique descriptive `info` field as the `msg=` to derive the "Index seed" for shuffling as well as each new share payload until a threshold of codex32 strings exist.
3. Additional derived shares are derived as per the BIP _but_ at shuffled indexes.
4. If necessary, the `master_seed` is recovered from `k` shares as per the BIP and the identifier is updated based on this fresh master seed's fingerprint.

```python3
def generate_shares(master_key="", user_entropy="", n=31, k="2", ident="NOID",
                    seed_length=16, existing_codex32_strings=None):
    """
    Generate new codex32 shares from provided or derived entropy.

    :param master_key: BIP32 extended private master key from bitcoind.
    :param user_entropy: User-provided entropy for improved security.
    :param n: Total number of codex32 shares to generate (default: 31).
    :param k: Threshold parameter (default: 2).
    :param ident: Identifier (4 bech32 characters) or 'NOID' (default).
    :param seed_length: Length of seed (16 to 64 bytes, default: 16).
    :param existing_codex32_strings: List of existing codex32 strings.
    :return: Tuple: master_seed (bytes), list of new codex32 shares.
    """
    master_seed = b""
    if existing_codex32_strings is None:
        existing_codex32_strings = []
    new_shares = []
    num_strings = len(existing_codex32_strings)
    if (not validate_codex32_string_list(existing_codex32_strings, False)
            and existing_codex32_strings):
        return None
    available_indices = list(CHARSET)
    for string in existing_codex32_strings:
        k, ident, share_index, payload = decode("ms", string)
        available_indices.remove(share_index)
        if share_index == "s":
            master_seed = payload
        seed_length = len(payload)

    if num_strings == int(k) and not master_seed:
        master_seed = recover_master_seed(existing_codex32_strings)
    if master_seed:
        master_key = BIP32Node.from_rootseed(master_seed, xtype="standard")
    elif master_key:
        master_key = BIP32Node.from_xkey(master_key)
    else:
        return None
    key_identifier = hash_160(master_key.eckey.get_public_key_bytes())
    entropy_header = (seed_length.to_bytes(length=1, byteorder="big")
                      + bytes("ms" + k + ident + "s", "utf") + key_identifier)
    salt = entropy_header + bytes(CHARSET[n] + user_entropy, "utf")
    # This is equivalent to hmac-sha512(b"Bitcoin seed", master_seed).
    password = master_key.eckey.get_secret_bytes() + master_key.chaincode
    # If scrypt absent visit OWASP Password Storage or use pbkdf2_hmac(
    # 'sha512', password, salt, iterations=210_000 * 64, dklen=128)
    derived_key = hashlib.scrypt(
        password, salt=salt, n=2 ** 20, r=8, p=1, maxmem=1025 ** 3, dklen=128)
    index_seed = hmac.digest(derived_key, b"Index seed", "sha512")[:32]
    available_indices.remove("s")
    available_indices = shuffle_indices(index_seed, available_indices)
    tmp_id = "temp" if ident == "NOID" else ident

    # Generate new shares, if necessary, to reach a threshold.
    for i in range(num_strings, int(k)):
        share_index = available_indices.pop()
        info = bytes("Share payload with index: " + share_index, "utf")
        payload = hmac.digest(derived_key, info, "sha512")[:seed_length]
        new_shares.append(encode("ms", k, tmp_id, share_index, payload))
    existing_codex32_strings.extend(new_shares)

    # Derive new shares using ms32_interpolate.
    for i in range(int(k), n):
        fresh_share_index = available_indices.pop()
        new_share = derive_share(existing_codex32_strings, fresh_share_index)
        new_shares.append(new_share)

    # Relabel the new shares with default ID.
    master_seed = recover_master_seed(existing_codex32_strings)
    if ident == "NOID":
        ident = "".join([CHARSET[d] for d in ms32_fingerprint(master_seed)])
        new_shares = relabel_codex32_strings("ms", new_shares, k, ident)

    return master_seed, new_shares
```

- **Notes on Key Derivation Function Selection:**

It is **critical** that the KDF used for the derived key and the identifier encryption be _computationally expensive_ for the hardware implementing this standard. Time costs between 2 and 10 seconds with a significant memory cost (relative to available memory) will offer the best protection for low entropy user input against grinding attacks, in acceptable user time.

**Scrypt** with parameters **n=2 ** 20, r=8, p=1** is used above. This requires 1 GB and a couple seconds on a modern laptop. If 1GB of memory is unavailable `pbkdf2_hmac('sha512', password, salt, iterations=210_000 * 64, dklen=128)` offers similar protection with slower runtime. Using [Argon2id](https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#argon2id) with a minimum configuration of 1024 MiB of memory, an iteration count of 4, and 4 degrees of parallelism can offer even better protection than Scrypt and may be implemented by wallets.

I do not prescribe any particular KDF but if using a different KDF or parameters than **scrypt** with **n=2 ** 20, r=8, p=1**, compliant wallet implementations MUST:

1.  Make the KDF function used and its cost parameters be publicly and permanently available for auditors.
2. Use **_at least_** 2048 PBKDF2 iterations of HMAC-SHA512 **OR** its equivalent.
    - See [OWASP Password Storage Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html) for the latest equivalencies.
4. Not use _any_ fixed cost KDF, hash based message authentication code or simple hash function in place of scrypt.

They also must generally:
4. Display the `app_entropy`, if any, unconditionally.
5. Allow the user to provide at least 128-bits of `user_entropy` (128+ character minimum, 1024 is a reasonable maximum.) AFTER displaying app entropy.

I welcome everyone's feedback on this proposal. #59

BenWestgate:
Closes #64
 
This is a draft of unshared secrets generation support.

**Summary:** Updated to discourage weird bitlengths, improve error correction handling. Added details on generation support for codex32 secrets.

Guidance for codex32 share sets generation can come in a later PR as it will depend largely this seed export guidance.

I covered **string display** and **confirmation**, from my experience designing these and receiving user feedback.

Confirm section presents two methods depending on keyboard available:
1) Full-verification method I used in Bails that is very easy and well liked by beta testers, and
2) Partial-verification method Blockstream used in Jade, borrowed near verbatim from: https://github.com/Blockstream/Jade/blob/690c8b1c5f9ccaf18954111a6ee8e124f2b75eec/main/process/mnemonic.c#L269-L398
  - Essentially, change only word-list and Jade can display and confirm 128-bit codex32 strings just as well as bip39.

Missing/TBD:
- BIP85 app (if it merges).
- Reference GUI
  - I have had for years a complete reference Gtk GUI of the "Full confirm" dialog, Bails uses it. I'll try to package that with the display and import views to make a PyPI `codex32-gui` I package for devs to play with.
- Reference CLI/Library
  - I wrote a [`codex32`](https://pypi.org/project/codex32/) PyPI library, my apologies if I was not supposed to take this name. The implementation is at least as clean as the rust in this repo and passes all BIP93 test vectors.

### DIFF
--- a/docs/wallets.md
+++ b/docs/wallets.md
@@ -37,7 +37,7 @@ If a wallet is unable to meet these specifications, it is not an ECW and it SHOU
 
 ## Import Support
 
-Wallets SHOULD support import of 128- and 256-bit seeds; other lengths are optional. 128-bit seeds encode as 48-character codex32 strings, including the `MS1` prefix. 256-bit seeds encode as 74-character codex32 strings. For other bit-lengths, see the BIP.
+Wallets SHOULD support import of 128- and 256-bit seeds; 512-bit seeds are optional; other lengths are NOT RECOMMENDED. 128-bit seeds encode as 48-character codex32 strings, including the `MS1` prefix. 256-bit seeds encode as 74-character codex32 strings. For other bit-lengths, see the BIP.
 
 The process for entering codex32 strings is:
 
@@ -45,34 +45,33 @@ The process for entering codex32 strings is:
    * The user MUST be able to enter all bech32 characters.
    * ECWs MUST also allow entry of `?` which indicates an erasure (unknown character).
    * The user SHOULD NOT be able to enter mixed-case characters.
-   * If the header is invalid, the wallet SHOULD highlight the problem and request confirmation from the user before allowing additional data to be entered.
-     * An invalid header is one that starts with a character other than `0` or `2` through `9`, or one which starts with `0` but whose share index is not `S`. For shares after the first, a header is also invalid if its threshold and identifier do not match those of the first share or whose share index matches any previous share.
-     * ECWs MAY replace the offending characters of the header with `?`.
+   * If the header is invalid, a the wallet SHOULD highlight the problem and if not an ECW, request confirmation from the user before allowing additional data to be entered.
+     * An invalid header is one that starts with a character other than `0` or `2` through `9`, or one which starts with `0` but whose share index is not `S`.
+     * ECWs SHOULD replace the offending characters of the header with `?`.
    * Wallets MAY:
-     * Allow users to enter invalid characters, at their discretion. (This may be useful to guide error correction, by attempting to replace commonly confused characters.)
+     * Allow users to enter invalid characters, at their discretion.
      * Use predictive text for on-screen keyboards to suggest the codex32 checksum characters but if so MUST require user to manually accept the prediction.
-     * Indicate when the entry has a valid checksum, e.g. by highlighting the string green or displaying the "Submit" option but they MUST NOT submit a string with a valid checksum without user request.
+     * Indicate when an entry has a valid checksum, e.g. by highlighting the string green or displaying the "Submit" option but they MUST NOT submit a string with a valid checksum without user request.
    * ECWs MAY additionally indicate when an entry of sufficient length to correct has an invalid checksum, e.g. by highlighting the string red or displaying an "Attempt Correction" option.
-
-
 1. Once the first string is fully entered, the wallet MUST validate the checksum and header before accepting it.
    * If the checksum does not pass, then an ECW:
       * MUST attempt error correction of substitution errors and erasures.
-      * MAY attempt correction by deleting and/or inserting characters, as long as the resulting string has a valid length for a codex32 string. ECWs MAY assume the correct length is the closest of 48 or 74.
+      * MAY attempt correction by deleting and/or inserting characters, as long as the resulting string has a valid length for a codex32 string. ECWs MAY assume the correct length is the closest of 48, 74, or 118.
       * MUST show any valid correction candidate found to the user for confirmation rather than silently applying it.
          * If insertion and/or deletion correction candidates are found, the shortest edit distance valid string SHOULD be displayed.
            * This is the sum of all edits with erasures and deletes weighted 1 and substitutions and insertions weighted 2.
          * ECWs displaying a candidate correction MAY highlight corrected 4-character windows and/or specific correction locations.
 1. After the first string has been entered and accepted, the wallet now knows the identifier, threshold value and valid length.
-   * If the first string had index `S`, this was the codex32 secret and the import process is complete.
-   * Otherwise, the fourth character of the share will be a numeric character between `2` and `9` inclusive. The user must enter this many shares in total.
-   * Wallets MAY encrypt and store recovery progress, to allow recovery without having all shares available at once. The details of this are currently outside of the scope of this specification.
+    * If the first string has index `S`, this is the codex32 secret and the import process is complete.
+    * Otherwise, the fourth character of the share will be a numeric character between `2` and `9` inclusive. The user must enter this many shares in total.
+    * Wallets MAY encrypt and store recovery progress, to allow recovery without having all shares available at once. The details of this are currently outside of the scope of this specification.
 1. The user should then enter the remaining shares, in the same manner as the first.
    * The wallet SHOULD pre-fill the header (threshold value and identifier).
-   * If the user tries to repeat an already-entered share index, they SHOULD be prevented from entering additional data until it is corrected.
-      * The wallet MAY guide the user by indicating that a share index has been repeated;
-      * ECWs may use `?` as a share index arbitrarily many times. If the user indicates they are not repeating the share, the share index SHOULD be replaced by `?`.
-   * If the checksum fails, the wallet MAY attempt correction by deleting and/or inserting characters. However, the wallet MUST assume the valid length of all subsequent shares is equal to the valid length of the first share, so the number of characters inserted and deleted must net out to the correct length.
+     * For shares after the first, the header is invalid if its threshold and identifier do not match those of the first share or whose share index matches any previous share.
+   * If the user tries to repeat an already-entered share index, they SHOULD be prevented from entering additional data until the user confirms they are not repeating a share.
+     * non-ECWs SHOULD highlight the error and request the user to correct it before allowing additional data to be entered.
+     * ECWs SHOULD replace the offending share index with `?`.
+    * If the checksum fails, the wallet MAY attempt correction by deleting and/or inserting characters. However, the wallet MUST assume the valid length of all subsequent shares is equal to the valid length of the first share, so the number of characters inserted and deleted must net out to the correct length.
 1. For all invalid codex32 strings entered, if an ECW is able to correct the errors (by deletion, insertion, substitution and/or filling erasures), it MUST show the corrected string to the user and request confirmation that the corrected string **exactly matches** the user's copy of the data. It MUST NOT silently apply corrections without approval from the user.
     * If no valid string is found with a correct hrp, header and unique index within correction distance limits or within 10 seconds of search, give up.
     * ECWs MAY warn the user they've repeated a share if the only valid string found exactly matches a previously entered share.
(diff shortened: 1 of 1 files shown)


BenWestgate:
@roconnor-blockstream This is ready for review.

roconnor-blockstream:
I don't know if I'm quite ready to completely abandon 160-bit and 192-bit seeds just yet.  Maybe I could be convinced.

BenWestgate:
> I don't know if I'm quite ready to completely abandon 160-bit and 192-bit seeds just yet. Maybe I could be convinced.

We should discourage lengths too close they prevent _**insert/delete**_-correcting ECWs from assuming the correct valid length.

Lets start with the BIP39 lengths:

image

A 128-bit seed with 6-8 erroneous insertions can be corrected easily by a desktop, IF the _**insert**_-correcting ECW can assume 128-bit length.  (removing 8 insertions has 1.6% chance of two solutions, 64% for 9.)

160-bit seeds MUST be discouraged as they can be mistaken for a 128-bit seed with a correctable amount of insertions.

A seed with 3 deletions can be corrected using your haskell impl to check every combo of 3 positions as erasures. (3 deletions has <6% chance of two solutions, ~100% for 4+.)

Now say we have an invalid but correctable length string, for software to know whether to try insertions or deletions, the next recommended length after 128-bits can be `128 + 8 * 5 + 3 * 5 = 183 < bitlength < 201 = 256 - 8 * 5 - 3 * 5`.

So 128, 192, 256 and 512 bitlengths are safe to recommended.
160, 224 and others must be discouraged since they complicate insert/delete ECWs assumptions about length.

BenWestgate:
As suggested in https://github.com/bitcoin/bips/pull/2040#discussion_r2561023486:

Lets define deterministic encoding defaults (identifier, padding, output share indices order) here and/or [here](https://github.com/bitcoin/bips/pull/1958) first.

I think encoders/wallets SHOULD:

1. select padding using a CRC-`w` where: `w = pad_bits_needed`, `poly = 1 << w | 3`, `init = 1`, `const = 1 << w - 1`, `refIn = false`, and `refOut = false`.
1. use the first 4 characters of the bech32-encoded bip32 masterkey fingerprint as the secret seed identifier if unspecified.
1. deterministically shuffle the _n_ share indices output.
    - Previously the bip85 app shuffled share indices but reviewers wanted it to output single strings so it was removed.

BenWestgate:
@roconnor-blockstream, as for 160-bit seeds:

I do not think "ms" secret seeds MUST NOT use them, I am saying "ms" secret seeds SHOULD use bitlengths divisible by 64 bits to simplify "insert/delete"-correcting ECWs (IDC-ECW) assumptions about valid lengths."

> [3](https://datatracker.ietf.org/doc/html/rfc2119#section-3). SHOULD   This word, or the adjective "RECOMMENDED", mean that there
>    may exist valid reasons in particular circumstances to ignore a
>    particular item, but the full implications must be understood and
>    carefully weighed before choosing a different course.

The implications of using 160-bits would not be the wallet rejects your valid secret seed. But rather that invalid 160-bit entries might receive 128-bit master seed correction suggestions from an IDC-ECW.

As I think "MS" IDC-ECWs MAY not check this length as their correction performance is maximized by a wider spacing.

This means 160-bit seeds are not interoperable with all IDC-ECW.

This RECOMMENDATION impacts the lengths our generators SHOULD generate.  As we don't want to produce seed lengths that harm ECW performance if they choose to support them.

Again valid reasons exist to deviate, both on generation and IDC-ECW assumptions about lengths but improving error correction seems more important than not discouraging 160-bit, when 192-bit can be used instead.

PR #74, opened 2026-01-19
python-codex32 reference impl

BenWestgate:
https://github.com/apoelstra/rust-codex32/pull/2#issuecomment-3622479406:

> @BenWestgate if you have a rewrite feel free to open a PR -- if it's not too hard to review I'm happy to take it in.
> 
> But I think the correct direction to rewrite in is one where we add a rust-bech32 dependency, use that for all the checksumming and encoding stuff, and then here we add (a) utility methods, and (b) constructors and accessors for the id/threshold/share index.

I have wrote this python implementation that covers more tests than the rust one does. It also does the approach above and relies on general checksum and bech32 modules. So it also passes Bech32/Bech32m tests.

It includes new test vectors for https://github.com/bitcoin/bips/pull/2040 and https://github.com/bitcoin/bips/pull/2077.

From: https://github.com/BenWestgate/python-codex32

Published on PyPI: https://pypi.org/project/codex32/

### DIFF
--- a/reference/python-codex32/LICENSE
+++ b/reference/python-codex32/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2026 Ben Westgate
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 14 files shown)


apoelstra:
I vote to merge this with a cursory review; later when I improve rust-codex32, I will steal all the test vectors from this implementation, and presumably if they pass on both implementations, they're both correct :P.

cursory review ACK 4857e187421dbff45301194d4bd8870c410d71f9 -- checked that this is shaped like it does codex32 stuff, that there's nothing obviously malicious, no filesystem or network access, etc.

cc @roconnor-blockstream can you let me know if I should hold off on merging this? Fine to ping me out-of-band.

BenWestgate:
I nACK https://github.com/BlockstreamResearch/codex32/commit/4857e187421dbff45301194d4bd8870c410d71f9 because it allows encoding shares from padded bytes. When the padding is not uniform, independent random it can leak up to `(k-1) * pad_len` bits of secret to attackers with `k-1` derived shares.

Comment explaining the problem: https://github.com/BlockstreamResearch/codex32/commit/4857e187421dbff45301194d4bd8870c410d71f9#r198130738

The rust impl allows this too, and an issue was opened where a user of our libraries was confused he could not recover the same secret bytes from derived share bytes.

image

The best solution is refuse to encode or decode shares from/to bytes. Especially now that https://github.com/bitcoin/bips/pull/2258 defines "S" decoding to be application specific.

I have some BIP0173 style test vectors in the https://github.com/bitcoin/bips/pull/2040#discussion_r2625828686 that are worth stealing as well.

BenWestgate:
that gentleman's vibe-coded implementation constructs/decodes shares from/to u5 only, like bip85 app-93 and my comments elsewhere.  It's always cool seeing AI put valid stray comments to use.

PR #76, opened 2026-08-20
rust: select checksum by expanded codeword length

### LINKED ISSUE #75: Change reference implementation to select checksum type "application agnostically"
For future applications, codex32 needs to be able to validate the header, string length, and select between regular/Long checksum to validate strings without caring what the human-readable prefix is, other than non-zero length printable characters.

To answer the question "Is this string valid in the codex32 format?"

@apoelstra suggested this issue here: https://github.com/bitcoin/bips/pull/2040#issuecomment-5267460865
> But I guess you should open a separate (small) PR to change the reference implementation so we can discuss it independent of anything else.

Is the reference I should change this one?
https://github.com/BlockstreamResearch/codex32/tree/master/reference/rust-codex32

To keep the PR small, I will only change the checksum selection math. I won't try to "generalize for any HRP" yet. That will be easiest to review.

BenWestgate:
Account for the HRP expansion when selecting and slicing regular and long checksums. Enforce the invalid 94–95 value gap and update the affected test vectors.

Aligns the reference implementation with bitcoin/bips#2258.

Closes https://github.com/BlockstreamResearch/codex32/issues/75

### DIFF
--- a/reference/rust-codex32/src/lib.rs
+++ b/reference/rust-codex32/src/lib.rs
@@ -142,6 +142,7 @@ impl Codex32String {
         // Compute the checksum
         checksum.input_hrp(hrp)?;
         checksum.input_data_str(real_string)?;
+        checksum.input_own_target();
         for ch in checksum.into_residue() {
             s.push(ch.to_char());
         }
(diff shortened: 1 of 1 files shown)


BenWestgate:
added the new test vectors from bip-0093.

apoelstra:
In 581a472eea2c2254a40b3c753dd35e12d916f9b7:

It's worth commenting that the "expanded length" function differs from the "expanded length" in the BIP, because the BIP includes the checksum symbols while this code does not. Hence comparisons against 81 instead of 94 e.g.

This also continues to enforce the 128-symbol maximum, which we should change to enforce the fixed set of allowed lengths.

(I know this PR predates the changes to the BIP PR, I'm just writing them out for the record.)

reference/rust-codex32/src/lib.rs
```diff
@@ -328,6 +337,12 @@ impl Codex32String {
         if id.len() != 4 {
             return Err(Error::IdNotLength4(id.len()));
         }
+        if share_idx != Fe32::S {
```
BenWestgate, line 340: the easy fix to constructing mispadded shares from bytes is have this command only work for "S"

BenWestgate (commented):
I also fixed the issue with creating shares from bytes.  Ready for re-review so the Rust matches bip93 bitcoin/bips master.

